PR #45, opened 2022-01-02
Fix overrun in decodeBLEJson loop and check for valid input data.

h2zero:
## Description:
Bugfix: the `decodeBLEJson` loop was using the wrong size of the devices array as the limiter.

If the input data does not contain valid "servicedata" or "manufacturerdata", return immediately rather than parsing all the devices. 

## Checklist:


1technophile:
Thanks!

PR #50, opened 2022-01-08
Add RuuviTag RAWv1 decoder

koenvervloesem:
## Description:

This adds a decoder for the [RuuviTag RAWv1](https://docs.ruuvi.com/communication/bluetooth-advertisements/data-format-3-rawv1) data format.

I added the [official test vector](https://docs.ruuvi.com/communication/bluetooth-advertisements/data-format-3-rawv1#case-valid-data) to the tests.

This test fails because I couldn't figure out how to handle negative values in the [temperature](https://docs.ruuvi.com/communication/bluetooth-advertisements/data-format-3-rawv1#temperature) and [acceleration](https://docs.ruuvi.com/communication/bluetooth-advertisements/data-format-3-rawv1#acceleration) fields. I'd like to get some help here.

## Checklist:


1technophile:
Thanks for the PR @koenvervloesem 

Could you try to replace for the acc: 

```
 "accx":{
         "decoder":["value_from_hex_data", "manufacturerdata", 16, 4, false, true],
         "post_proc":["/", 1000]
      },
```
By

```
"accx":{
         "decoder":["value_from_hex_data", "manufacturerdata", 16, 4, false],
         "post_proc":["/", 1000]
      },
```
I'm suspecting a bug in the negative value handling when explicitely defined as true

koenvervloesem:
I tried this, but it doesn't make any difference, both with the test vector and realtime sensor values.

koenvervloesem:
Ok, the acceleration values now work with the fix in https://github.com/theengs/decoder/pull/51. However, the temperature field isn't using two's complement representation of the signed integers, but it's using a sign bit: MSB is the sign bit, and the next 7 bits are the decimal value. How can this be implemented in Theengs Decoder?

1technophile:
Maybe by using a condition on this MSD bit, here is an example: https://github.com/theengs/decoder/blob/3808f5635b60f5c43401570376227caadb5250d2/src/devices/iBeacon_json.h#L26

You will have to specify 2 times your property `tempc` and `_tempc`

Depending on the MSD value you apply a different post processor

koenvervloesem:
I thought the condition applied to a property is only able to test for a fixed value, isn't it? At least that's what all the examples show and how I read its documentation and the source code. Here I would have to test for one bit of the byte value, and I don't think that's currently possible?

1technophile:
@h2zero do you think we can extract the bit so as to sign the rest of the value?

h2zero:
I'll see what I can think of for this case.

h2zero:
The encoding format for the temp on this will require a new post processing function. I will look into adding this but it may take some time.

h2zero:
Just to update this, I have created a new function for handling BCF format data. I have tested it with this PR and it is passing. I will open a PR for this shortly.

h2zero:
@koenvervloesem I have opened the mentioned PR, in order to support this device using that commit you will need to change the device definition to the following:
```
const char* _RuuviTag_RAWv1_json = "{\"brand\":\"Ruuvi\",\"model\":\"RuuviTag\",\"model_id\":\"RuuviTag\",\"condition\":[\"manufacturerdata\",\"=\",32,\"index\",0,\"990403\"],\"properties\":{\"hum\":{\"decoder\":[\"value_from_hex_data\",\"manufacturerdata\",6,2,false,false],\"post_proc\":[\"/\",2]},\"tempc\":{\"decoder\":[\"bf_value_from_hex_data\",\"manufacturerdata\",8,4,false,true]},\"pres\":{\"decoder\":[\"value_from_hex_data\",\"manufacturerdata\",12,4,false,false],\"post_proc\":[\"+\",50000,\"/\",100]},\"accx\":{\"decoder\":[\"value_from_hex_data\",\"manufacturerdata\",16,4,false,true],\"post_proc\":[\"/\",1000]},\"accy\":{\"decoder\":[\"value_from_hex_data\",\"manufacturerdata\",20,4,false,true],\"post_proc\":[\"/\",1000]},\"accz\":{\"decoder\":[\"value_from_hex_data\",\"manufacturerdata\",24,4,false,true],\"post_proc\":[\"/\",1000]},\"volt\":{\"decoder\":[\"value_from_hex_data\",\"manufacturerdata\",28,4,false,false],\"post_proc\":[\"/\",1000]}}}";
/*R""""(
{
   "brand":"Ruuvi",
   "model":"RuuviTag",
   "model_id":"RuuviTag",
   "condition":["manufacturerdata", "=", 32, "index", 0, "990403"],
   "properties":{
      "hum":{
         "decoder":["value_from_hex_data", "manufacturerdata", 6, 2, false, false],
         "post_proc":["/", 2]
      },
      "tempc":{
         "decoder":["bf_value_from_hex_data", "manufacturerdata", 8, 4, false, true]
      },
      "pres":{
         "decoder":["value_from_hex_data", "manufacturerdata", 12, 4, false, false],
         "post_proc":["+", 50000, "/", 100]
      },
      "accx":{
         "decoder":["value_from_hex_data", "manufacturerdata", 16, 4, false, true],
         "post_proc":["/", 1000]
      },
      "accy":{
         "decoder":["value_from_hex_data", "manufacturerdata", 20, 4, false, true],
         "post_proc":["/", 1000]
      },
      "accz":{
         "decoder":["value_from_hex_data", "manufacturerdata", 24, 4, false, true],
         "post_proc":["/", 1000]
      },
      "volt":{
         "decoder":["value_from_hex_data", "manufacturerdata", 28, 4, false, false],
         "post_proc":["/", 1000]
      }
   }
})"""";*/

const char* _RuuviTag_RAWv1_json_props = "{\"properties\":{\"hum\":{\"unit\":\"%\",\"name\":\"humidity\"},\"tempc\":{\"unit\":\"°C\",\"name\":\"temperature\"},\"pres\":{\"unit\":\"hPa\",\"name\":\"pressure\"},\"accx\":{\"unit\":\"g\",\"name\":\"accelerationX\"},\"accy\":{\"unit\":\"g\",\"name\":\"accelerationY\"},\"accz\":{\"unit\":\"g\",\"name\":\"accelerationZ\"},\"volt\":{\"unit\":\"V\",\"name\":\"voltage\"}}}";
/*R""""(
{
   "properties":{
      "hum":{
         "unit":"%",
         "name":"humidity"
      },
      "tempc":{
         "unit":"°C",
         "name":"temperature"
      },
      "pres":{
         "unit":"hPa",
         "name":"pressure"
      },
      "accx":{
         "unit":"g",
         "name":"accelerationX"
      },
      "accy":{
         "unit":"g",
         "name":"accelerationY"
      },
      "accz":{
         "unit":"g",
         "name":"accelerationZ"
      },
      "volt":{
         "unit":"V",
         "name":"voltage"
      }
   }
})"""";*/
```

koenvervloesem:
@h2zero Thanks! This works now indeed.

1technophile:
Thanks for the PR, could you add your device documentation [here ](https://github.com/theengs/decoder/tree/development/docs/devices) please.
Also you will need to add the page [here](https://github.com/theengs/decoder/blob/3171cc6c25870ab173ef47253caf4db1cadcb341/docs/.vuepress/config.js#L31)

koenvervloesem:
Sure.

Just a question. I've added the RuuviTag RAWv2 data format too in https://github.com/theengs/decoder/pull/60. These are both for the same hardware, but these are data formats for different firmware versions.

Should I document both formats as separate devices? And should I use a separate model and model ID for both formats? Or should I use the same model and model ID and one documentation page for the RuuviTag? And in the latter case, to which device header file should I link?

1technophile:
I would document it separately and use a different model_id and different pages for the moment. In the same principle of the LYWSD03MMC.
We may introduce later on a new attribute for firmware and have to merge those versions.

koenvervloesem:
Done, and I've done the same for the other PR.

1technophile:
Thanks!

PR #52, opened 2022-01-09
Fix XMTZC04HM condition

1technophile:
## Description:
the previous condition was too restrictive and didn't take into account some servicedata case

## Checklist:


1technophile:
Need to investigate further, this fix doesn't seems to correct the issue

PR #67, opened 2022-02-12
Universal XMTZC04HM decoder

DigiH:
XMTZC04HM decoder with uuid condition only

Unless there were other reasons why an additional condition of **"servicedata", "index", 0,** was added to the MiScale v1 decoder, only using the uuid, similar to the Miscale_v2 decoder, will allow for other XMTZC04HM variants to be recognised.

Related to

https://github.com/1technophile/OpenMQTTGateway/issues/1166

## Checklist:

### DIFF
--- a/src/devices/XMTZC04HM_json.h
+++ b/src/devices/XMTZC04HM_json.h
@@ -2,13 +2,13 @@
 
 // A condition is done on the first character = 2 so as to verify if the weight is stabilized, the decoder doesn't decode weights that are not stabilized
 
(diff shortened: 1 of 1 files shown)


kgarrels:
ok, let's continue here.

Now also unstabilized weights are sent. 
The decoder should check if byte 0 is either '2' or 'a', I think that will yield stabilized values, no matter if you still stand on the scale or not ("stabilized" and "weight removed" bits).

'a' should actually be good enough, if the BLE scan happens while the weight is still on the scale, we can just wait for the next round when the weight is gone.

I did the modification in my local copy of the decoder, and it seems to work.

This the correct data format:
https://github.com/oliexdev/openScale/wiki/Xiaomi-Bluetooth-Mi-Scale

DigiH:
So the additional condition check "servicedata", "index", 0, "2" was for stabilised weight, which now makes sense to me. Introducing the recognition of additional unstabilised readings probably ins't the best idea though.

@kgarrels It might be best to look into why your scale didn't seem to stabilise and then send the service data accordingly - possibly it standing on uneven ground, carpet, movement on the scale?!?

Looking back into the issue and PR history, which I really should have done before, this 'stabilised' check was specifically introduced because of unwanted "undefined" values.

https://github.com/1technophile/OpenMQTTGateway/issues/801

https://github.com/1technophile/OpenMQTTGateway/pull/868

Closing this PR as NA for removing the additional "servicedata", "index", 0 completely.

We can have further discussion about additional checks in the initial issue, possibly in conjunction to the Mi Body composition scale issue

https://github.com/1technophile/OpenMQTTGateway/issues/760

PR #77, opened 2022-02-23
Update XMTZC05HM to include XMTZC02HM

DigiH:
Having confirmed that the first version of the Mi Body Composition Scale XMTZC02HM is also being decoded by the second versions's  XMTZC05HM decoder I included and described both in the docs and decoder and updated the tests.

This should finalise these two Mi Scale models both being identified and recognised, additionally differentiated by their different names in the JSON. 

XMTZC02HM/XMTZC05HM update in docs, decoder and tests

## Checklist:


DigiH:
@h2zero @1technophile , sorry to bother you again with some query, hopefully the last time with the Mi Scale decoders.

Having implemented the current XMTZC05HM for my scale in my openHAB I found that when weighing an **object** or a **person** not barefoot, when no **impedance** is reported in the JSON my manual thing/item definition throws an warning because on the non-existent JSONPATH transformation $.impedance for these broadcasts. I've rejigged my code to avoid these warning messages, but since I haven't got any experience with HA discovery or even openHAB auto discovery, could this cause similar issues there?

So I thought that instead reporting impedance in such cases as 0 would be the best option, and tried both

```
      "_impedance":{
         "condition":["servicedata", 3, "4"],
         "decoder":["static_value", "0"]
      }
```

and

```
      "_impedance":{
         "condition":["servicedata", 3, "4"],
         "decoder":["static_value", 0]
      }
```

https://github.com/DigiH/decoder/blob/impedancetest/src/devices/XMTZC05HM_json.h

but it is always reported in the JSON as a string in quotes, compared to an actual impedance reading as a number.

`…"model_id":"XMTZC02HM/XMTZC05HM","weighing_mode":"object","unit":"kg","weight":2.5,"impedance":"0"}`

In my manual setup this is not an issue, but could this be an issue with discovery setups?

Would be good to hear your thoughts on this. While I would prefer no impedance in the JSOn at all when not properly registered, it really also depends how it best works with discovery.

Thanks

h2zero:
That is odd, I'm unsure why it would report the static value in quotes when it is an integer. 

The code that handles this is pretty basic:
```
jsondata[sanitizeJsonKey(kv.key().c_str())] = decoder[1];
```

Could be that we need to add a test here for `is<int>()` and set the value with `as<int>()`

DigiH:
I'm not really worried about it, if it doesn't mess with discovery in any way, that is, if a **0** impedance is actually necessary instead of not publishing it all all when it's not available.

But might be something to look into if other static numbers crop up in future decoders. Thanks

1technophile:
Thanks!

PR #93, opened 2022-03-13
ON HOLD - Govee H5055 temp/battery only

DigiH:
As discussed in

https://github.com/theengs/decoder/issues/92

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] I accept the [DCO](https://github.com/theengs/decoder/blob/development/docs/participate/development.md#developer-certificate-of-origin).

PR #97, opened 2022-03-20
Initial NOT implementation

DigiH:
This ist just as much a test for myself to delve deeper into the decoder code, as the implementation of the previously discussed NOT condition testing. Maybe not implemented as elegantly as could be, so any comments on this are welcome ;)

This allows for some previous long stringed OR conditions to be greatly reduced, as in the MiBand act_bpm property for example, and also allows for some unrealistic temperatures not to be published when the probes are not connected as in the Inkbird 6XS BBQ and Tenergy SOLIS 6 BBQ cases - amended the test cases accordingly. I assume the Inkbird 2X BBQ and Inkbird 4XS BBQ would benefit from the same implementation, but didn't want to change the decoders, as the current test cases don't confirm this assumption.

## Checklist:

### DIFF
--- a/docs/participate/adding-decoders.md
+++ b/docs/participate/adding-decoders.md
@@ -82,6 +82,22 @@ Here we have a single property that defines a value that we want to decode. The 
 The second parameter is the index of the data source to look for the value. The third parameter is the value to test for.
 If the condition is met the data will be decoded and added to the JsonObject.
 
+Property conditions also allow for a NOT comparison, as in
+```
+ "properties":{
+      "tempc":{
+         "condition":["manufacturerdata", 24, "!", "ffff"],
+         "decoder":["value_from_hex_data", "manufacturerdata", 24, 4, true, false],
+         "post_proc":["/", 10]
+      },
+```
+
+where then the fourth parameter is the value to test for.
+
+::: warning Note
+The NOT comparison is case sensitive! Therefor any NOT comparisons should be defined in lower case, as this is the format in which devices' "servicedata" and "manufacturerdata" are being reported.
+:::
+
 `decoder` is a JSON array that specifies the decoder function and parameters to decode the value.
 The first parameter is the name of the function to call, The available functions are:
 - "value_from_hex_data"  - converts the hex string value into an `integer` or `double` type.
--- a/src/decoder.cpp
+++ b/src/decoder.cpp
@@ -300,14 +300,27 @@ int TheengsDecoder::decodeBLEJson(JsonObject& jsondata) {
         bool cond_met = prop_condition.isNull();
 
         for (int i = 0; i < cond_size; i += 4) {
-          if (svc_data &&
-              strstr((const char*)prop_condition[i], "servicedata") != nullptr &&
-              svc_data[prop_condition[i + 1].as<int>()] == *prop_condition[i + 2].as<const char*>()) {
-            cond_met = true;
-          } else if (mfg_data &&
-                     strstr((const char*)prop_condition[i], "manufacturerdata") != nullptr &&
-                     mfg_data[prop_condition[i + 1].as<int>()] == *prop_condition[i + 2].as<const char*>()) {
-            cond_met = true;
+          if (strstr((const char*)prop_condition[i + 2], "!") != nullptr) {
+            if (svc_data &&
+                strstr((const char*)prop_condition[i], "servicedata") != nullptr &&
+                svc_data[prop_condition[i + 1].as<int>()] != *prop_condition[i + 3].as<const char*>()) {
+              cond_met = true;
+            } else if (mfg_data &&
+                      strstr((const char*)prop_condition[i], "manufacturerdata") != nullptr &&
+                      mfg_data[prop_condition[i + 1].as<int>()] != *prop_condition[i + 3].as<const char*>()) {
+              cond_met = true;
+            }
+            i++;
+          } else {
+            if (svc_data &&
+                strstr((const char*)prop_condition[i], "servicedata") != nullptr &&
+                svc_data[prop_condition[i + 1].as<int>()] == *prop_condition[i + 2].as<const char*>()) {
+              cond_met = true;
+            } else if (mfg_data &&
+                      strstr((const char*)prop_condition[i], "manufacturerdata") != nullptr &&
+                      mfg_data[prop_condition[i + 1].as<int>()] == *prop_condition[i + 2].as<const char*>()) {
+              cond_met = true;
+            }
           }
 
           if (!cond_met && cond_size > (i + 3) && *prop_condition[i + 3].as<const char*>() == '|') {
--- a/src/devices/IBT_2X_json.h
+++ b/src/devices/IBT_2X_json.h
@@ -7,20 +7,22 @@ const char* _IBT_2X_json = "{\"brand\":\"Inkbird\",\"model\":\"iBBQ\",\"model_id
    "condition":["name", "index", 0, "iBBQ","&","manufacturerdata", ">=", 24, "index", 0, "0000","&","manufacturerdata", "<=", 28, "index", 0, "0000"],
    "properties":{
       "tempc":{
-         "condition":["manufacturerdata", 0, "00000000"],
+         "condition":["manufacturerdata", 0, "00000000", "&", "manufacturerdata", 16, "!", "f6ff"],
          "decoder":["value_from_hex_data", "manufacturerdata", 16, 4, true, false],
          "post_proc":["/", 10]
       },
       "_tempc":{
+         "condition":["manufacturerdata", 20, "!", "f6ff"],
          "decoder":["value_from_hex_data", "manufacturerdata", 20, 4, true, false],
          "post_proc":["/", 10]
       },
       "tempc2":{
-         "condition":["manufacturerdata", 0, "00000000"],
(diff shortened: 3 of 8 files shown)


1technophile:
Thanks @DigiH ! Could you update the docs also?

DigiH:
@1technophile sure, but just came across an issue with the new condition. It seems that the upper/lower case of the (literal) NOT comparison matters. Now this is with the test cases being different for different devices, but is it also the same with actual devices, or are servicedata and manufacturerdata always reported the same (lower case I assume?) In that case we should actually change all the affected test cases, as otherwise the tests might work, but an actual device will fail.

Just confirmed with the latest failed test run.

1technophile:
>  always reported the same (lower case I assume?)

Yes they are always lower case, and I think we should consider this as a prerequisite in the decoder input.

>  I assume the Inkbird 2X BBQ and Inkbird 4XS BBQ would benefit from the same implementation, but didn't want to change the decoders, as the current test cases don't confirm this assumption.

I confirm the assumption with my 4XS

DigiH:
> Yes they are always lower case, and I think we should consider this as a prerequisite in the decoder input.

Yes, that would also flag test cases with upper case data provided, not that it seems to matter for decoding, but with the conditions.

> I confirm the assumption with my 4XS

Thanks :) implemented it as well for the 4XS and the 2X.

Let me update the docs in this commit as well …

DigiH:
Done

h2zero:
Looks good to me

h2zero (approved):

DigiH:
Just also added a NOTE to the documentation about the case-sesitivity.

1technophile:
Thanks!

DigiH:
Hi @1technophile @h2zero I think my implementation wasn't quite 100% correct enough to also work properly with the 'nested' property condition testing.

While trying to implement the NOT condition into the still on hold **H5055-tempsonly** decoder I found that with adding

`, "&", "manufacturerdata", 14, "!", "ffff"`

at the end the conditions ALWAYS get validated, ignoring the ORs before.

https://github.com/DigiH/decoder/commit/c8757cac164b3597857db621e92e11d1bb1bf0b2

resulting in non-sensical test failure

```
1: Found : 24 {"brand":"Govee","model":"BluetoothBBQThermometer","model_id":"H5055","tempc1":23,"tempf1":73.4,"tempc2":115,"tempf2":239,"tempc3":23,"tempf3":73.4,"tempc4":115,"tempf4":239,"tempc5":23,"tempf5":73.4,"tempc6":115,"tempf6":239,"batt":70}
1: Key:value count mismatch, result 16, expected 8
1: Expected: {"brand":"Govee","model":"Bluetooth BBQ Thermometer","model_id":"H5055","tempc1":23,"tempf1":73.4,"tempc2":115,"tempf2":239,"batt":70}
1: Got JSON: {"brand":"Govee","model":"BluetoothBBQThermometer","model_id":"H5055","tempc1":23,"tempf1":73.4,"tempc2":115,"tempf2":239,"tempc3":23,"tempf3":73.4,"tempc4":115,"tempf4":239,"tempc5":23,"tempf5":73.4,"tempc6":115,"tempf6":239,"batt":70}
1/2 Test #1: run_test_ble .....................***Failed    0.04 sec
```

This also didn't show up in the initial submission, where I also implemented a 

`, "&", "manufacturerdata", 16, "!", "f6ff"`

in the IBT_2X decoder, as it's only with the AND but no OR conditions before?!?

I've also restored my not-implementation branch and would appreciate if you have any pointers in figuring out the issue at hand.

Or were the nested OR plus AND conditions only implemented in the model condition and shouldn't currently be used in property conditions? Just a bit confused right now :)

Thanks

DigiH:
I think the IBT_2X decoder cases would actually need to be switched around, with the 

`"condition":["manufacturerdata", 0, "00000000"],`

being for the _ versions, as otherwise

```
      "_tempc2":{
         "decoder":["value_from_hex_data", "manufacturerdata", 24, 4, true, false],
         "post_proc":["/", 10]
      }
```

will not work on the shorter only 24 long manufacturerdata. But we can also now use

`"condition":["manufacturerdata", 0, "!", "00000000"],`

for the shorter manufacturerdata version once the remaining nested issue is sorted out.

DigiH:
Messed up with the decoders for 2X BBQ and Inkbird 4XS BBQ, seemingliy forgetting to run my script for the const char conversion for them. Now with them corrected I get

https://github.com/DigiH/decoder/commit/5b82bc7b5ba11481acf520e69e6df3c13346aa15

looking like the NOT comparison only evaluates just the first hex digit in the comparison, with "fa00" for example for IBT-4XS, resulting in

```
1: Found : 13 {"brand":"Inkbird","model":"iBBQ","model_id":"IBT-4XS","tempc":26,"tempf":78.8,"tempc2":26,"tempf2":78.8}
1: Key:value count mismatch, result 7, expected 11
1: Expected: {"brand":"Inkbird","model":"iBBQ","model_id":"IBT-4XS","tempc":26,"tempf":78.8,"tempc2":26,"tempf2":78.8,"tempc3":25,"tempf3":77,"tempc4":25,"tempf4":77}
1: Got JSON: {"brand":"Inkbird","model":"iBBQ","model_id":"IBT-4XS","tempc":26,"tempf":78.8,"tempc2":26,"tempf2":78.8}
1/2 Test #1: run_test_ble .....................***Failed    0.03 sec
```

Could this 'first digit' only evaluation happen with 
`
mfg_data[prop_condition[i + 1].as<int>()] != *prop_condition[i + 3].as<const char*>()`

and is something similar happening already with

` mfg_data[prop_condition[i + 1].as<int>()] == *prop_condition[i + 2].as<const char*>()`

as the IBT_2X test cases and results still confuse me? 

```
"tempc2":{
         "condition":["manufacturerdata", 0, "00000000", "&", "manufacturerdata", 20, "!", "f6ff"],
         "decoder":["value_from_hex_data", "manufacturerdata", 20, 4, true, false],
         "post_proc":["/", 10]
      },
```

should not really kick in for the first shorter variant of the IBT_2X version, but it clearly does (also only evaluating the first '0' !?!), as otherwise the variant
```
      "_tempc2":{
         "condition":["manufacturerdata", 24, "!", "f6ff"],
         "decoder":["value_from_hex_data", "manufacturerdata", 24, 4, true, false],
         "post_proc":["/", 10]
      }
```

would throw an exception with only 24 long manufacturerdata!?! Ignoring the recently added "manufacturerdata", 20, "!", "f6ff" for this.

So that the current NOT but already previous multi-digit property comparisons don't/didn't work?!?

h2zero:
@DigiH, I'll have a look into this tonight. Do you have capability to run the tests locally with a debugger?

DigiH:
@h2zero I've been _playing_ around a bit and currently at

https://github.com/DigiH/decoder/commit/a86c5ee7818188a4fae8e00da38a2595df777cb1

to see how property conditions can have just a single hex digit, as well as longer hex 'strings' as in the model conditions. Still all very new to me, also with the local debugger side :) 

Thanks for also having a look into it.

**P.S.** with the above I think I got it to work, but please do check my syntax and thinking behind it when you get the time ;) and by work I mean condition checking for single as well as longer hex 'strings' if NOT combined with other 'nested' conditions, as the H5055-tempsonly decoder is still failing with

https://github.com/DigiH/decoder/commit/c8757cac164b3597857db621e92e11d1bb1bf0b2

and I also had to remove the initially submitted multiple _IBT_2X_json conditions to single NOT or already existing positive conditions, which also work now

https://github.com/DigiH/decoder/commit/e0f5e6624bf9ea2d7a0bedba74a3fadfb81d05fb

Leaving everything as is for the time being, until you further comment on it.

Thanks

h2zero:
I've gone over this in the testing using your changes. I have a couple thoughts I will test, will update tomorrow.

h2zero:
@DigiH I have created a new branch here: https://github.com/theengs/decoder/tree/check-prop-cond-func

This uses your code from the above commits but adds a new method as a prequel to other future changes to the logic for easier maintenance. This is starting to get larger than originally considered, separating out segments of the decode function is the logical next step.

Please test that branch and let me know your thoughts.

DigiH:
Thanks a lot @h2zero

> adds a new method as a prequel to other future changes to the logic for easier maintenance. This is starting to get larger than originally considered, separating out segments of the decode function is the logical next step.

for restructuring the code for better future proofing and clearer structure. Everything looks and works fine, I only had to add an IBT_4XS decoder fix and also added an extra test case to account and monitor for its new NOT functionality as well.

https://github.com/DigiH/decoder/commits/h2zero-new-check-prop-cond-func

If it's ok I can create a PR with the above, or you can merge your decoder changes and I will submit the IBT_4XS changes separately, as not to mix up commits.

**CORRECTION**: not all is working fine, as commented in

https://github.com/theengs/decoder/pull/93

with the multiple ORs before the AND NOT condition the NOT fails to be recognised.

**Full disclosure** - I've also already been working on a binary-bitwise condition checking in decoder.cpp, still with the code before your restructuring, and it is working fine already, but still very messy code. While I don't really want to inflate the code with lots of new decoding options I think this one is very crucial, as it greatly reduces most (if not all) existing multiple OR checkings to one single binary bit check, eliminating a great deal of such 'nested' OR AND conditions.

@h2zero @1technophile Would be good to know how you feel about this - trying to fix the 'nested' issue currently showing in the H5055 decoder, or possibly also adding the binary-bit decoder functionality to try and avoid future nested OR AND  as much as possible, or both? ;)

PR #101, opened 2022-03-26
Inkbird fixes and 6xS merge with Solis 6

DigiH:
• Inkbird leading manufacturerdata fixes
• merge of 6XS with Solis 6
• test corrections
• docs fixes

## Checklist:


src/devices/IBT_6XS_SOLIS6_json.h
```diff
@@ -0,0 +1,70 @@
...
+/*R""""(
+{
+   "brand":"Inkbird/Tenergy",
+   "model":"iBBQ/SOLIS6",
+   "model_id":"IBT-6X(S)/SOLIS-6",
```
1technophile: I'm worried about the parenthesis in the model id. Not sure how the different client programs handle this. Maybe we stick to IBT-6XS/SOLIS-6 ?

1technophile:
Thanks!

src/devices/IBT_6XS_SOLIS6_json.h
DigiH: Yes, I wasn't actually sure about that, so best to keep that just in the docs with the parenthesis.

Also thinking about renaming the Solis 6 to just Solis, as it is also being sold with just 2 probes, but still has all the 6 connectors for additional probes. What do you think? Or keep it with the 6 as it still has 6 connectors ;)

DigiH: Resubmitted with just the parenthesis change.

1technophile: Yes, makes sense to rename it to Solis only

PR #118, opened 2022-05-21
Consistent quotes

DigiH:
Consistent post_proc and AND OR quotes

While working on the SwitchBot Meter decoder I realised that "post_proc" examples and my implementation in the S1 decoder are inconsistent with single ' and double quotation marks.

While both seem to work fine I think it would be good to have consistent double quotes as generally used everywhere else.

## Checklist:


h2zero:
Good catch, single quotes are invalid in JSON but ArduinoJSON lets us get away with it.

PR #124, opened 2022-05-25
bit_static_value decoder function

DigiH:
New "bit_static_value" decoder function

Implementation in SwitchBot devices, Xiaomi scales and CGH1 decoders.

Updated docs.

## Checklist:

### DIFF
--- a/src/decoder.cpp
+++ b/src/decoder.cpp
@@ -46,6 +46,10 @@ typedef double (TheengsDecoder::*decoder_function)(const char* data_str,
                                                    int offset, int data_length,
                                                    bool reverse, bool canBeNegative);
 
+typedef double (TheengsDecoder::*staticbitdecoder_function)(const char* data_str,
+                                                            const char* source_str, int offset, int bitindex,
+                                                            const char* falseresult, const char* trueresult);
+
 /*
  * @brief Revert the string data 2 by 2 to get the correct endianness
  */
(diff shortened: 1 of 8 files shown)


src/decoder.cpp
```diff
@@ -149,6 +153,18 @@ int TheengsDecoder::data_length_is_valid(size_t data_len, size_t default_min,
...
 }
 
+uint8_t TheengsDecoder::getBinaryData(char ch) {
+  uint8_t data = 0;
+  if (ch >= '0' && ch <= '9')
+    data = ch - '0';
+  else if (ch >= 'a' && ch <= 'f')
+    data = 10 + (ch - 'a');
+  else
+    return false;
```
h2zero: This else case isn't needed since the returned data will be 0 anyway.

src/decoder.cpp
```diff
@@ -149,6 +153,18 @@ int TheengsDecoder::data_length_is_valid(size_t data_len, size_t default_min,
...
+
+return data;
```
h2zero: Please indent.

h2zero (commented):
Minor nitpicks, looks good otherwise.

src/decoder.cpp
DigiH: Done, thanks

src/decoder.cpp
DigiH: Done, thanks

1technophile:
This enables a good cleanup, thanks

PR #127, opened 2022-06-17
b-parasite soil sensor

DigiH:
Thanks to @mjeanrichard I was made aware of this open hardware/open source soil moisture and ambient temperature/humidity/light sensor. In the spirit of open source this could be a desirable alternative to the Xiaomi/VegTrug Mi Flora (sans fertility) devices for some users.

https://github.com/rbaron/b-parasite

Not testet with any actual device - totally hydroculture here for all indoor plants - but verified with other implemented test cases for all hardware revisions.

## Checklist:

### DIFF
--- a/src/devices/BPARASITE_json.h
+++ b/src/devices/BPARASITE_json.h
@@ -1,0 +1,64 @@
+const char* _BPARASITE_json = "{\"brand\":\"rbaron\",\"model\":\"b-parasite\",\"model_id\":\"BPv1.0-1.2\",\"condition\":[\"servicedata\",\">=\",32,\"index\",0,\"1\",\"|\",\"servicedata\",\">=\",32,\"index\",0,\"2\",\"&\",\"uuid\",\"index\",0,\"181a\"],\"properties\":{\"tempc\":{\"condition\":[\"servicedata\",0,\"1\"],\"decoder\":[\"value_from_hex_data\",\"servicedata\",8,4,false,true],\"post_proc\":[\"/\",1000]},\"_tempc\":{\"condition\":[\"servicedata\",0,\"2\"],\"decoder\":[\"value_from_hex_data\",\"servicedata\",8,4,false,true],\"post_proc\":[\"/\",100]},\"hum\":{\"decoder\":[\"value_from_hex_data\",\"servicedata\",12,4,false,false],\"post_proc\":[\"/\",655.35]},\"moi\":{\"decoder\":[\"value_from_hex_data\",\"servicedata\",16,4,false,false],\"post_proc\":[\"/\",655.35]},\"lux\":{\"condition\":[\"servicedata\",1,\"bit\",0,1],\"decoder\":[\"value_from_hex_data\",\"servicedata\",32,4,false,false]},\"volt\":{\"decoder\":[\"value_from_hex_data\",\"servicedata\",4,4,false,false],\"post_proc\":[\"/\",1000]}}}";
+
+/* R""""(
+{
+   "brand":"rbaron",
+   "model":"b-parasite",
+   "model_id":"BPv1.0-1.2",
+   "condition":["servicedata", ">=", 32, "index", 0, "1", "|", "servicedata", ">=", 32, "index", 0, "2", "&", "uuid", "index", 0, "181a"],
+   "properties":{
+      "tempc":{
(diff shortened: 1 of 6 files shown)


src/devices/BPARASITE_json.h
```diff
@@ -0,0 +1,64 @@
+const char* _BPARASITE_json = "{\"brand\":\"rbaron\",\"model\":\"b-parasite\",\"model_id\":\"BPv1.0-1.2\",\"condition\":[\"servicedata\",\">=\",32,\"&\",\"uuid\",\"index\",0,\"181a\"],\"properties\":{\"tempc\":{\"condition\":[\"servicedata\",0,\"1\"],\"decoder\":[\"value_from_hex_data\",\"servicedata\",8,4,false,true],\"post_proc\":[\"/\",1000]},\"_tempc\":{\"condition\":[\"servicedata\",0,\"2\"],\"decoder\":[\"value_from_hex_data\",\"servicedata\",8,4,false,true],\"post_proc\":[\"/\",100]},\"hum\":{\"decoder\":[\"value_from_hex_data\",\"servicedata\",12,4,false,false],\"post_proc\":[\"/\",655.35]},\"moi\":{\"decoder\":[\"value_from_hex_data\",\"servicedata\",16,4,false,false],\"post_proc\":[\"/\",655.35]},\"lux\":{\"condition\":[\"servicedata\",1,\"bit\",0,1],\"decoder\":[\"value_from_hex_data\",\"servicedata\",32,4,false,false]},\"volt\":{\"decoder\":[\"value_from_hex_data\",\"servicedata\",4,4,false,false],\"post_proc\":[\"/\",1000]}}}";
+
+/* R""""(
+{
+   "brand":"rbaron",
+   "model":"b-parasite",
+   "model_id":"BPv1.0-1.2",
+   "condition":["servicedata", ">=", 32, "&", "uuid", "index", 0, "181a"],
```
rbaron: Thanks for this!

How does this condition work here? The length of the service data should be either 16 (for protocol v1) or 18 (for protocol v2). The description is [here](https://github.com/rbaron/b-parasite/tree/main/code/b-parasite#bluetooth-low-energy-advertisement-data-encoding).

I think it would be more future proof we add a condition for "protocol version 1 or 2".

DigiH: Thanks for having a look @rbaron, the whole servicedata length (32 digits for the 16 bytes) in the model condition together with the AND of the uuid is only for distinguishing the b-parasite from other possible BLE broadcasting device in the vicinity. Also it will still recognise any future updated protocols, as long as they are >= 32. Alternatively we could OR - "servicedata", "=", 32 OR "servicedata", "=", 36, but the uuid is still the deciding factor in the model condition. 

> I think it would be more future proof we add a condition for "protocol version 1 or 2".

Further down in the property conditions we use the protocol version (1 and 2 so far) for the correct temperature decoding and the **has_lux** bit for only then registering the luminance. The same protocol check or additional bit checks can be done for any properties which might need them in the future.

**How about this?**

`"condition":["servicedata", ">=", 32, "index", 0, "1", "|", "servicedata", ">=", 32, "index", 0, "2", "&", "uuid", "index", 0, "181a"],`

Restricting the current decoder to only protocol 1 and 2, until it will be updated for and future protocol implementations.

rbaron: Your last suggestion is indeed is what I had in mind. Exaggerating example: if in v3 we actually remove some sensors or change how they are encoded, we may break old parsers, so it's better to explicitly condition on the versions.

Tks again!

rbaron: PS.: We just need to double check the precedence of those operators in this implementation. Example: is it (v1 || v2) && uuid == 181a (correct) or is it v1 || (v2 && uuid == 181a) (incorrect).

DigiH: It's (v1 || v2) && uuid == 181a, how we need it. Similar, but even more complex example here

https://github.com/theengs/decoder/blob/development/src/devices/XMTZC04HM_json.h#L7

rbaron: Awesome! 👌

1technophile:
Thanks @DigiH @rbaron @mjeanrichard ! Nice add

PR #133, opened 2022-06-29
Fix TX power unit of iBeacon: dBm instead of int

koenvervloesem:
## Description:

This fixes the unit of the TX power property of the iBeacon decoder: it should be dBm, not int. I noticed this while looking at Theengs Explorer's output, which shows the property's unit.

## Checklist:

### DIFF
--- a/src/devices/iBeacon_json.h
+++ b/src/devices/iBeacon_json.h
@@ -31,7 +31,7 @@ const char* _ibeacon_json = "{\"brand\":\"GENERIC\",\"model\":\"iBeacon\",\"mode
    }
 })"""";*/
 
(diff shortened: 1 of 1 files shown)


DigiH:
Nice catch!

Am also still wondering if we should include "@1m", possibly in the unit as well as the docs, or if this is just the protocol definitions, but not all 'iBeacons' stick to it, although actual varying power would be a bit strange.

koenvervloesem:
Good point. To be correct, this should probably be:

```json
      "power":{
         "unit":"dBm",
         "name":"tx power @ 1 m"
      }
```

DigiH:
Maybe something for the discussions first. I'm only aware of the iBeacon protocol devices which allow for this `tx power @ 1m` to be set by the user for refined distance calculation. Ruuvi tag also has

```
"tx":{
         "unit":"dBm",
         "name":"TX power"
      }, 
```

but not having one I'm unsure about if having the same settability/functionality. And while we're at it best to bring in line the upper/lower case spelling of all as well ;)

koenvervloesem:
For RuuviTag (I have those and I wrote the decoder) it's the real TX power, not the one measured at 1 meter: https://docs.ruuvi.com/communication/bluetooth-advertisements/data-format-5-rawv2

PR #134, opened 2022-06-29
Python build version

DigiH:
Not sure if we're really too concerned about this, but until Decoder might become a submodule of  Explorer I was just a bit surprised when looking at my pip3 packages

```
…
TheengsDecoder                 0.1.0
TheengsGateway                 0.2.0
…
```

Much more reassuring now with

```
…
TheengsDecoder                 0.4.0
TheengsGateway                 0.2.0
…
```

even if I built the dev branch :)

## Checklist:


1technophile:
Too many version definition  ! 
Thanks for pointing this.

PR #136, opened 2022-06-30
iBeacon-txpower

DigiH:
Changing the iBeacon decoder `power` to the more appropriate `txpower`, in line with other devices already reporting and Decoder registering their `txpower`, and for distinction between actual Watt Power of the iNode decoder for example,. Additional clarification of its @ 1 m implementation in the unit name and docs.

## Checklist:


1technophile:
Thanks !

Issue #137, opened 2022-07-01
Support service data for multiple UUIDs


koenvervloesem:
**Is your feature request related to a problem? Please describe.**

I have a sensor device that advertises service data for two UUIDs, and I want to create a decoder for it (https://github.com/theengs/decoder/pull/138). Currently, Theengs Decoder seems to support only one `servicedatauuid` and one `servicedata` property.

**Describe the solution you'd like**

This is how service data for my particular device are sent to the decoder now:

```python
❯ python3 examples/python/ScanAndDecode.py |grep RDL
D8:4B:C1:98:69:1E RSSI: -71 AdvertisementData(local_name='RDL52832', manufacturer_data={76: b'\x02\x15\xfd\xa5\x06\x93\xa4\xe2O\xb1\xaf\xcf\xc6\xeb\x07dx%\x00\x01\x00\x02\xd8'}, service_data={'00001803-0000-1000-8000-00805f9b34fb': b'\xd8K\xc1\x98i\x1e\x00\x01\x00\x02\x07\x03\xe8d', '00000318-0000-1000-8000-00805f9b34fb': b'\x15O8E\x01\x00\x00\x01\x00\x00\x00\x00\x00\x00\t\x07'})
data sent to decoder:  {"servicedatauuid": "1803", "servicedata": "d84bc198691e000100020703e864", "manufacturerdata": "4c000215fda50693a4e24fb1afcfc6eb0764782500010002d8", "name": "RDL52832"}
```

The example Python script only sends the first service data to the decoder, in this case for UUID 1803, while for this particular device I'm interested in the service data for UUID 0318.

What I would like is that I could send something like the following JSON data to the decoder and the service data for both UUIDs should then be available for further decoding:

```json
{"servicedata": {"1803": "d84bc198691e000100020703e864", "0318": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"}, "manufacturerdata": "4c000215fda50693a4e24fb1afcfc6eb0764782500010002d8", "name": "RDL52832"}
```

**Describe alternatives you've considered**

I could first check the name of the device (e.g. in Theengs Explorer) and then send the service data for UUID 0318 to the decoder if that specific device is detected, but this is just a hack.

**Additional context**

I'm not sure there are many devices that advertise service data on multiple UUIDs (this is the first one I discovered), but a general approach for this case seems nice to have.

1technophile:
[...]

PR #138, opened 2022-07-02
Add RDL52832 decoder

koenvervloesem:
## Description:

This adds a decoder for the RDL52832 sensor beacon. It's an iBeacon that also advertises temperature, humidity and accelerometer readings.

There are still some issues before this can be merged:

- ~~The device is only correctly detected if the decoder gets the service data for UUID 0318 (see https://github.com/theengs/decoder/issues/137). I have a test script at the bottom of this description that does this.~~
- The acceleration properties are not correctly computed. I define a `.cal` value three times to extract Xt, Yt and Zt (see the data format below) and use these values, but it seems the decoder is only using the latest `.cal` value, so `accx`, `accy` and `accz` are calculated based on the same extracted value (Zt). See https://github.com/theengs/decoder/issues/139 for more information.

## Data format

The data format looks like this (see https://github.com/mike-rankin/ESP32_iBeacon_Sensor/blob/main/Code/iBeacon_Test.ino):

```c
// Data format for the RadioLand RDL52832 iBeacon
//
// service UUID 0x0318:
//    0    1    2    3    4    5    6    7    8    9   10   11   12   13   14   15
// +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
// | Th | Tl | Hh | Hl | Xs | Xw | Xt | Xh | Ys | Yw | Yt | Yh | Zs | Zw | Zt | Zh |
// +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
// Temperature (T), stored as a 16-bit value equal to Celcius * 256
// Humidity (H), stored as a 16-bit value equal to Humidity % * 256
// Accelerometer Axes: (X,Y,Z)
//    s = sign (1=negative, 0=positive)
//    w = whole value (0 or 1)
//    t = tenths (0-9)
//    h = hundredths (0-9)
```

## Checklist:

## Test script

```python
import asyncio
import json
import struct
from bleak import BleakScanner
from TheengsDecoder import decodeBLE as dble
from TheengsDecoder import getProperties, getAttribute

def detection_callback(device, advertisement_data):
    data_json = {}

    if device.name == "RDL52832":

        print(device.address, "RSSI:", device.rssi, advertisement_data)
        if advertisement_data.service_data['00000318-0000-1000-8000-00805f9b34fb']:
            dstr = '00000318-0000-1000-8000-00805f9b34fb'
            # TheengsDecoder only accepts 16 bit uuid's, this converts the 128 bit uuid to 16 bit.
            data_json['servicedatauuid'] = dstr[4:8]
            dstr = advertisement_data.service_data['00000318-0000-1000-8000-00805f9b34fb'].hex()
            data_json['servicedata'] = dstr

        if advertisement_data.manufacturer_data:
            dstr = str(struct.pack('<H', list(advertisement_data.manufacturer_data.keys())[0]).hex())
            dstr += str(list(advertisement_data.manufacturer_data.values())[0].hex())
            data_json['manufacturerdata'] = dstr

        if advertisement_data.local_name:
            data_json['name'] = advertisement_data.local_name

        if data_json:
            print("data sent to decoder: ", json.dumps(data_json))
            data_json = dble(json.dumps(data_json))
            print("TheengsDecoder found device:", data_json)

            if data_json:
                dev = json.loads(data_json)
                print(getProperties(dev['model_id']))
                brand = getAttribute(dev['model_id'], 'brand')
                model = getAttribute(dev['model_id'], 'model')
                print("brand:", brand, ", model:", model)


async def main():
    scanner = BleakScanner()
    scanner.register_detection_callback(detection_callback)
    await scanner.start()
    await asyncio.sleep(5.0)
    await scanner.stop()

    for d in scanner.discovered_devices:
        print(d)

asyncio.run(main())
```


h2zero:
Tests pass if you prepend an underscore to each of the .cal values after the first, just like the other items.

DigiH:
So I ordered one as well now, I just hope they send the correct one ;) @koenvervloesem, from the data sheet (only found the Chinese version so far) it seems it has exactly the same writing commands for setting its name, uuid, major, minor and txpower as my nrf51822 beacons. With seemingly a slight difference for setting the broadcast interval and an additional password setting option.

How is your experience with the above?

DigiH:
Also from my search on Aliexpress all the nrf52832 iBeacons seem to only come in one version, with the default txpower transmission, compared to the nrfnrf51822, which specifically come in two different version with either txpower or battery voltage encoded in byte 24.

Therefore I think we can safely remove the volt property to slim down the decoder at least a tiny bit.

koenvervloesem:
> Tests pass if you prepend an underscore to each of the .cal values after the first, just like the other items.

That's a hell of an undocumented feature :-) Cool, this works indeed. I have added some extra test data with the device in various orientations.

> Therefore I think we can safely remove the volt property to slim down the decoder at least a tiny bit.

I removed the property.

I think the decoder is ready now.

h2zero:
>That's a hell of an undocumented feature :-) 

Lol, yes it is. Had a look at the code this morning and thought, "well why wouldn't it work?", and so it does 😄 .

koenvervloesem:
> How is your experience with the above?

Actually I haven't changed the default settings yet. I was more interested in the sensor readings :-)

PR #140, opened 2022-07-04
TH2-P01B extension

DigiH:
Merging of P01B into TH2 decoder, and extension of TH2 to include humidity which is also reported for at least some hardware revisions of it.

## Checklist:


DigiH:
The question remaining is, should we get them all together with TH1 as well? ;)

Ok, maybe not. After going through all the manuals for

TH1
TH1 Mini
TH1 PLUS

and

TH2 (Temp only)
TH2 (Temp & Humidity)
TH2 PLUS

I think we should keep them separately, as the TH2, apart from its PLUS version, are IPX4 water splash proof.

With the TH2 PLUS, TH1 and TH1 PLUS also allowing for the external probes we might have to add options to the decoder if and when required.

1technophile:
Thanks, yes I think we can keep it separate for the moment.

PR #141, opened 2022-07-04
iNode Energy Meter extension

DigiH:
@1technophile while unfortunately we haven't heard back from the users, the actual electricity power readings are the same as with the decoder before, but with the additional 

• allow all manufacturerdata, index 1 variations to be recognised for decoding
• added the m³ water meter functionality
• added current average and sum units depending on electricity or water metering
• included lowBattery key, only when true

options. Not having any regressions, this is only really an improvement, especially with the above listed first bullet point, which failed the recognition with the previous decoder.

## Checklist:


1technophile:
Thanks for the improvement

PR #147, opened 2022-07-08
Add service UUID to condition of LYWSD02 decoder

koenvervloesem:
## Description:

This adds a check of the service data UUID to the LYWSD02 decoder, as suggested in https://github.com/theengs/decoder/pull/146.

## Checklist:


DigiH:
Thanks

I think this might also be a good idea for many other Xiaomi decoders, to narrow down the model conditions with servicedata UUID and solely the product id.

@koenvervloesem @1technophile do you have any other of these devices to confirm this suggestion?

koenvervloesem:
I have a Mi Flora, I'll investigate later.

PR #148, opened 2022-07-08
Narrow down conditions for Mi Flora decoder

koenvervloesem:
## Description:

This narrows down the conditions for the Mi Flora decoder, as suggested in https://github.com/theengs/decoder/pull/147. I don't have the model with product ID bc03 to check, but for my own devices it works and I adapted the tests to take into account the UUID too.

## Checklist:


DigiH:
For this case is the VegTrug just a rebranded Xiaomi, so that both have the "fe95" servicedata uuid, or could it have a different one?

I just realised that even in the test cases we don't have the "bc03" specific ID in the VegTrug servicedata, and if entered there the current model condition probably fails as the pseudo-netsted way of working would expect the AND to be after the initial OR conditions - I think ;)

Could you see how it runs with a corrected VegTrug servicedata?

DigiH:
> For this case is the VegTrug just a rebranded Xiaomi, so that both have the "fe95" servicedata uuid, or could it have a different one?

Apparently from what I found, yes, it has the same servicedata uuid. So that's that out of the way.

koenvervloesem:
If I change the product ID of the VegTrug test data, it still passes the tests.

DigiH:
> If I change the product ID of the VegTrug test data, it still passes the tests.

Yes, but because this current model condition

`"condition":["uuid", "index", 0, "fe95", "&", "servicedata", "index", 4, "9800", "|", "servicedata", "index", 4, "bc03"],`

acts like

`"condition":[["uuid", "index", 0, "fe95", "&", "servicedata", "index", 4, "9800"] "|", "servicedata", "index", 4, "bc03"],`

therefor only evaluating `"servicedata", "index", 4, "bc03"` for the VegTrug, ignoring the uuid completely. It should still pass the test even if you change the test uuid to "xxxx" - I hope ;)

I know, a bit special and still sometimes confusing to everybody involved in the initial implementation. Please have a look at my recent clarification docs PR

https://github.com/theengs/decoder/commit/bea99d36ac2d17715d539fa5cbc6a7070242568a

koenvervloesem:
Right, moved the UUID check to the end now.

PR #110, opened 2022-05-07
IBT2X(S) fix

DigiH:
Inkbird IBT2X(S) fix to included the S replacable battery model idetifying as "xBBQ"

https://community.openmqttgateway.com/t/only-openmqtt-sensors-appear-in-home-assistant-missing-bluetooth-devices/1892/9

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] I accept the [DCO](https://github.com/theengs/decoder/blob/development/docs/participate/development.md#developer-certificate-of-origin).

PR #154, opened 2022-07-18
IBT-2XS rogue high temp fix

DigiH:
Fix for

https://community.openmqttgateway.com/

While I initially thought it was a good idea to shorten the manufacturerdata comparison from `"index", 0, "00000000" ` to `"index", 2, "000000"` in 

https://github.com/theengs/decoder/pull/110

to include both 2X and 2XS versions in one decoder, it wasn't then clear that the IBT-2XS also sends **02** prefixed manufacturerdata of weirdly unrealistic high temperatures every now and then :)

We might want to wait just a bit before merging just to see if the other user also has the same **02** occurrences.

## Checklist:


DigiH:
I think we can actually merge this, as it's very unlikely that the other user's issues isn't the same **02** prefixed manufacturerdata, and even it it was still happening with the now defined

`… "manufacturerdata", "=", 28, "index", 0, "00000000", "|","manufacturerdata", "=", 28, "index", 0, "01000000"`

there wouldn't be much we could do about it on the decoder side.

What do you say?

1technophile:
I say let's go!

PR #182, opened 2022-08-15
cidc property introduction

DigiH:
Introduction of the cidc (company ID compliant) property, only required when **false** - companies not adhering to the Bluetooth SIG convention for the mfid.

## Checklist:


1technophile (approved):

1technophile:
Thanks

h2zero:
If I saw this earlier I would have suggested a simple code change to allow for `{"cidc":false}` in the JSON root which would assume true unless otherwise stated in JSON. Then all devices would have `{"cidc":true}` except those that have specified it, just like the docs in this change suggested. 

This would reduce some repetition and be more efficient IMO.

PR #185, opened 2022-08-18
Extract more data from ThermoBeacon advertisements

koenvervloesem:
## Description:

Adds properties for timestamp, maximum temperature, maximum temperature timestamp, minimum temperature, and minimum temperature timestamp to the ThermoBeacon decoder. These properties are extracted from the manufacturer data advertisements with length 44.

## Checklist:


DigiH:
I really like the fact that this is now also fetching the minimum and maximum temperatures, but the time stamps are a bit cryptic. 

Are the min/max temperatures the min and max of the whole time a ThermoBeacon has been active - as is suggested from the last test case with ~60 hours difference of the max and min time stamps, that is if the time stamps are uptime in seconds. Or how are they defined?

I think this information needs be added in the doc to allow for users to actually make useful calculations for when the max/min temperatures occurred, like
`datetime(now) - (time-time_max)seconds`
or such.

koenvervloesem:
Yes these timestamps seem to be uptime in seconds. They're all reset to 0 after removing and reattaching the battery, and they start counting up once a second.

As for the period the min and max temperatures are computed for, I didn't find any information about it. I also couldn't find a function to reset the time period to compute min and max in the Sensor Blue app. So for the time being, I would suppose that these are min and mix during the whole uptime.

PR #191, opened 2022-08-28
Fix device match conditions when OR cases used with data length.

h2zero:
## Description:
Changes the way data length checking works to account for validation of OR cases when the data length can match in multiple ways.

## Checklist:


DigiH:
Thanks so much, this is working fine with the two affected decoders in
https://github.com/DigiH/decoder/commit/ffe75577c430896dfe0d4302a7d358b5ad00f98d

Looks good to go then from my side.

PR #243, opened 2022-12-15
Fixed format string for size_t-typed variable

matthias-bs:
## Description:
%zu seems to be the correct format string for size_t-typed variables; size_t implementation is architecture-specific, e.g. unsigned int or long unsigned int.

## Checklist:
  - [x ] The pull request is done against the latest development branch


1technophile:
Good catch!

PR #253, opened 2023-01-07
Don't use deprecated method register_detection_callback

koenvervloesem:
## Description:

Changes the deprecated method `register_detection_callback` of the `BleakScanner` class in the Python example.

## Checklist:


DigiH:
Thanks

PR #281, opened 2023-02-05
Add id and rssi keys to data sent to decoder in Python example

koenvervloesem:
## Description:

Add id and rssi keys to data sent to decoder in Python example to be consistent with Theengs Gateway implementation.

## Checklist:


DigiH:
Thanks!

PR #319, opened 2023-03-25
Mi Band/Amazfit restriction

DigiH:
Mi Band/Amazfit decoder restriction to not catch wrongly decoded devices like the Amazfit Bip S and similar.

## Checklist:


toomyem:
Unfortunately, I wasn't able to decode any sensible data in `manufacturerdata` for BIP S watch. It looks like the data is somehow encoded, or the coding is not as simple as plain bytes. Every reading gives a completely different data set, ex:

```
570100 480967361d0078b09608774ca8df13e0 02c70cbd1794a2
570100 b6905f1ad520bcaa079ce2c5b5c69425 02c70cbd1794a2
570100 48ec7b6fbae4d3c9d99499a37a3e59f2 02c70cbd1794a2
570100 3835d0bf8780f1419a593ea7d626da7b 02c70cbd1794a2
570100 7985c46b89c31e4c35ae1170fd6cc0b4 02c70cbd1794a2
570100 dcdde701d61acdc010c59c77fad0bf8e 02c70cbd1794a2
```

So maybe, for now, just keep reading the 'steps' measurement, and omit the 'bpm' part.

DigiH:
> So maybe, for now, just keep reading the 'steps' measurement, and omit the 'bpm' part.

Thanks for the info @toomyem, so are the steps still broadcast in the servicedata with the same encoding for the Bip S? I haven't caught any Bip S servicedata braoadcasts here so far. It also requires active scanning for the service data to be received.

```
570100 480967361d0078b09608774ca8df13e0 02c70cbd1794a2
570100 b6905f1ad520bcaa079ce2c5b5c69425 02c70cbd1794a2
570100 48ec7b6fbae4d3c9d99499a37a3e59f2 02c70cbd1794a2
570100 3835d0bf8780f1419a593ea7d626da7b 02c70cbd1794a2
570100 7985c46b89c31e4c35ae1170fd6cc0b4 02c70cbd1794a2
570100 dcdde701d61acdc010c59c77fad0bf8e 02c70cbd1794a2
```

With this manufacturerdata, was this taken when an activity was actually started on the Bip S, and with activity heart rate sharing turned on for the Bip S in the Zepp Life App?

image

toomyem:
Similar data is being sent regardless if watch is in idle or any activity is started.

I also have a similar option in Zepp application to enable data sharing in real time, but I cannot turn it on. It apparently requires some third party app to be registered first.

DigiH:
> I also have a similar option in Zepp application to enable data sharing in real time, but I cannot turn it on. It apparently requires some third party app to be registered first.

So you do not see any steps with your Amazfit Bip S at all, i.e. no servicedata being received if you set [pubadvdata to true](https://docs.openmqttgateway.com/use/ble.html#advanced-advertisement-and-advanced-data-default-false)?

If you do, would you mind sharing the servicedata and its servicedatauuid?

toomyem:
Service data is being sent and it contains steps measurement. The missing part is heart rate.

"servicedata": "ef050000", -> 0x05ef -> 1519 steps (and it corresponds to actual reading on the watch).

What is interesting, I can connect to the device via BLE Tester app on my phone and then get heartbeats readings when some activity on the watch is started, and it also matches the actual readings on the watch:

obraz

EDIT:

It looks like ongoing activity is not required. That data is sent even if the watch is in idle, but less frequently.

DigiH:
As we're not able to connect to the Bip S to get the activity heart rate, and only can get it though the advertising manufacturerdata for the other models, I have changed the decoder to still recognised the Bip S as being part of the 
```
"brand":"Xiaomi/Amazfit",
"model":"Mi Band/Smart Watch",
"model_id":"MB/SW",
…
```
decoder , but only retrieve its steps counts, and restricted the activity heart rate retrieval from the manufacturerdata only for compatible 570102 models.

Thanks for your testing and input on this!

PR #327, opened 2023-04-06
Don't remove servicedata, tags and so on in Python _decoder.cpp

koenvervloesem:
## Description:

Don't remove servicedata, tags and so on in the `decode_BLE` function of the Python decoder's `_decoder.cpp`. This shifts the responsibility to remove unneeded information to the user of the library, such as Theengs Gateway.

## Checklist:


DigiH:
Thanks

PR #336, opened 2023-05-16
Export C functions for other languages

kabili207:
## Description:
This change exports the necessary decoder functions in C-style so they can be called by other languages. It's heavily based on the Python version already included.

Tested successfully with C# and Go

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,12 +24,17 @@ else()
 
     add_library(decoder
                 src/decoder.cpp
+                src/decoder_c.cpp
                 )
+
+    set_target_properties(decoder PROPERTIES
+        PUBLIC_HEADER shared/theengs.h)
 
     target_include_directories(decoder
         PUBLIC
             $<INSTALL_INTERFACE:arduino_json>
             $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/arduino_json/src>
+            ${CMAKE_CURRENT_SOURCE_DIR}/include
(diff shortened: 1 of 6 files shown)


koenvervloesem:
This is interesting. Is it also possible to include sample code for C# and/or Go, like the Python example in the examples directory?

kabili207:
I've added an example for Go, as well as squashed and rebased my commits.

I won't have access to the windows machine for a few days, so a proper C# example will have to wait, but the function declarations below should be the bare minimum to work with the library. A proper C# example would wrap these inside a class but, as I said, no access to windows at the moment.

```csharp
[DllImport("decoder.dll")]
public static extern IntPtr Theengs_NewDecoder();

[DllImport("decoder.dll")]
public static extern void Theengs_DestroyDecoder(IntPtr pointer);

[DllImport("decoder.dll")]
public static extern string Theengs_DecodeBLE(IntPtr pointer, string json_data);

[DllImport("decoder.dll")]
public static extern string Theengs_GetProperties(IntPtr pointer, string model_id);

[DllImport("decoder.dll")]
public static extern string Theengs_GetAttribute(IntPtr pointer, string model_id, string attribute);
```

1technophile:
That's great thanks!

DigiH:
Thanks @kabili207 !

Do you want to wait for being back on your Windows machine to add a C# example example to this PR, or do you want to do this later with a separate PR?

kabili207:
I can make a separate pull request for the C# example

PR #343, opened 2023-05-28
Make MS-CDP decoder model-only

koenvervloesem:
## Description:

Remove the properties of the MS-CDP decoder and make it a model-only decoder, as the devices can't be differentiated individually.

## Checklist:


DigiH:
Thanks

PR #346, opened 2023-06-03
Fix SBBT-002C decoder to detect new firmware

koenvervloesem:
## Description:

Starting from firmware 20230530-112223/v1.0.7-rc1@879855cd the ShellyBLU Button1 advertises the trigger-based device flag in the [BTHome device information](https://bthome.io/format/), so advertisements start with 44 instead of 40. This PR changes the decoder to detect both prefixes.

This PR also fixes the tag, as the decoder needs active scanning to identify the device type from its name.

## Checklist:


DigiH:
Thanks

PR #397, opened 2023-08-01
Automatic versioning from Git metadata

### LINKED ISSUE #392: Automatic versioning to easily build development versions from Git
**Is your feature request related to a problem? Please describe.**

Currently a development version of the Theengs Decoder Python package can't be easily built because setup.py doesn't have a valid version string. You have to temporarily change this to a valid version string, build the package and then don't forget to restore setup.py again.

**Describe the solution you'd like**

Automatic versioning from Git tags and commits, as implemented by [PyScaffold](https://github.com/pyscaffold/pyscaffold). This means that every commit can be built without having to manually change a version string.

**Describe alternatives you've considered**

None

**Additional context**

This would make it easier for users to test new features in development versions, such as in this case: https://github.com/1technophile/OpenMQTTGateway/issues/1555#issuecomment-1627734529

koenvervloesem:
## Description:

Uses [setuptools-scm](https://pypi.org/project/setuptools-scm/) to extract Python package version from Git metadata instead of declaring them manually. This allows building the package from every commit without having to manually change a version string, making it easier to test development versions. Fixes https://github.com/theengs/decoder/issues/392.

## Checklist:


DigiH:
Thanks

PR #409, opened 2023-08-21
Fix ABN07 decoder for firmware 4.0.1

koenvervloesem:
## Description:

The original firmware of April Brother's ABN07 had a bug so it advertised object IDs in the wrong order, which isn't compliant with the BTHome specification. After I notified April Brother about this, they fixed this behavior in firmware 4.0.1. This PR fixes Theengs Decoder to use the new format, and documents that users need to have firmware 4.0.1 or higher on their device.

## Checklist:


DigiH:
Thanks

PR #441, opened 2023-10-30
update batt LYWSDCGQ

Mit4el:
## Checklist:
  - [ ] The pull request is done against the latest development branch
  - [ ] I accept the [DCO](https://github.com/theengs/decoder/blob/development/docs/participate/development.md#developer-certificate-of-origin).


DigiH:
Hi @Mit4el 

This looks interesting, but could we ask you for some sample data where you think this would be required?

Are you using OpenMQTtGateway with Decoder or Theengs Gateway? 
Could you turn on Advertising and Advanced data on either and post some sample MQTT messages where this would be applicable?

As the battery property only uses a single 8-bit octet the first boolean defined endianness for "value_from_hex_data" it should not really matter if it is set to true or false - the test results come out all the same with either true or false.

If you can show us some sample data we might see that possibly some masking could be required for the battery level property.

DigiH:
@Mit4el 

Have a look at this change and see if it might not address whatever you have been seeing on your end

https://github.com/theengs/decoder/commit/55c4a669a0bd2e81367d660fe3cb303b210f5eae

Mit4el:
I use the decoder library in the open modular project "IoTmanager". Perhaps my changes are not required. After them, these data are received. The battery comes less frequently and separately from the rest of the data. Maybe I just didn't get it from. And before that, I used an old version of the decoder library where there was no battery. I need to test more, but I don't have the sensor, I wrote a module for using your library.

Data with my edits:
 {"id":"58:2D:34:3B:47:11","rssi":-76,"servicedata":"5020aa01ff11473b342d580a100164","serv
icedatauuid":"0xfe95","brand":"Xiaomi","model":"Mi Jia round","model_id":"LYWSDCGQ","type":"THB","batt":100,"
mac":"58:2D:34:3B:47:11"}

DigiH:
As stated above, your change does not actually make any difference to the result, as the endiannness is disregarded for a single 8-bit byte, it would only be confusing to set it to true.

You should receive the same battery data as is.

As you said, you might have missed it before with also using an older version of Decoder.

Closing this PR as not applicable.

PR #480, opened 2023-12-23
[CI] Upload Python package as artifact

koenvervloesem:
## Description:

Uploads the Python package built from a PR as an artifact, useful when we need a user to test a change. This doesn't require the user to set up a development environment. We can just let the user download the Python package from the artifacts in a GitHub Action run such as https://github.com/theengs/decoder/actions/runs/7309688303 and let them run `pip install <PACKAGE_FILE>`.

## Checklist:


DigiH:
Great addition for test build distribution! Thanks

PR #494, opened 2024-01-15
Add device decoder: TicWatch GTH & GTH Pro (GTH+).

jgbreezer:
First attempt at any decoder for a device. Please check carefully/make suggestions for the rest of it, and help me do a decent job, perhaps update docs for unclear things I mention below, and see my comments:
 - Invented a (unique) model_id.
 - Condition checks name starts with TicWatch GTH (Pro version adds '+'). Name suffixed by a space and last 4 characters of MAC).
 - Properties: mac decoder from manufacturer data (offset 4 hex digits, ie 0000<mac> in usual printed order)
 - Tag: picked 11(Body) type (vs.16). Not an RMAC. Believe continuous/active scanning may be best (how to tell? TheengsDecoder Gateway updates mqtt with rssi value regularly as is; is not cidc compliant, reports Ericsson AB mfr but Mobvoi/Ticwatch no known connection to them).
 - (Assumed "byte[0]" from adding-decoders doc is lsb not first digits in value. Is hex representation ok?

## Checklist:

### DIFF
--- a/src/devices/TicWatchGTH_json.h
+++ b/src/devices/TicWatchGTH_json.h
@@ -4,7 +4,7 @@ R""""(
    "model":"GTH/GTH Pro",
    "model_id":"TICWATCHGTH",
    "condition":["name", "index", 0, "TicWatch GTH"],
-   "tag": "0x0f0b",
+   "tag": "0f0b",
    "properties":{
       "mac":{
          "decoder":["value_from_hex_data", "manufacturerdata", 4]
--- a/src/devices/TicWatchGTH_json.h
+++ b/src/devices/TicWatchGTH_json.h
@@ -1,0 +1,13 @@
+R""""(
+{
+   "brand":"TicWatch",
+   "model":"GTH/GTH Pro",
+   "model_id":"TICWATCHGTH",
+   "condition":["name", "index", 0, "TicWatch GTH"],
+   "tag": "0x0f0b",
+   "properties":{
+      "mac":{
+         "decoder":["value_from_hex_data", "manufacturerdata", 4]
+      },
+   }
+})"""",


jgbreezer:
The existing `devices/*_json.h` files all have a C declaration with the python-like `R"""..."""` text (as used in the docs for an example), commented out. Is there an automated way to convert to that? I got stuck a bit into the python/README.md as I don't have ArduinoJSON.h so build failed and I don't see where to get it from. I'm trying to add support so I can use it in the Theengs Gateway (not using OMG, not thinking I need to) for a HomeAssistant setup.

jgbreezer:
MQTT post for the ticwatch, replaced the MAC by aabbccddeeff for privacy reasons (assuming the last 4 hex digits aren't the same for all of these devices) - this is for the GTH+ but I also previously owned the GTH (non-pro version).
{
  "manufacturerdata": "0000aabbccddeeff",
  "name": "TicWatch GTH+ EEFF",
  "id": "AA:BB:CC:DD:EE:FF",
  "rssi": _<NEGATIVE_INTEGER>_,
  "mfr": "Ericsson AB"
}

DigiH:
Looks like a good start @jgbreezer 

Just off to bed here, but will look at it more and comment tomorrow, but

> The existing devices/*_json.h files all have a C declaration with the python-like R"""...""" text (as used in the docs for an example), commented out. Is there an automated way to convert to that?

The commented out pretty JSON is for clarity and making changes. If you are using Visual Studio Code as well you can do a 

Screenshot 2024-01-16 at 00 43 47

from the Command Palette to convert it. And as the name suggests, it would also be auto-applied with a git commit.

koenvervloesem:
@jgbreezer Did you follow https://decoder.theengs.io/use/python.html#installing-a-development-version? ArduinoJSON should then be found, as the repository is cloned recursively with all submodules this way. If you have cloned the repository without its submodules, you can always get it later with `git submodule update --recursive`.

For the automatic conversion, it's expected that you do your changes in the comment with the R-string, and then automatically convert it to the C char array. To do this, install pre-commit with `pip install pre-commit` and then set up the Git pre-commit hooks with `pre-commit install` in the repository. After this, every `git commit` will automatically run the conversion script, and it's also available from Visual Studio Code as @DigiH showed.

DigiH:
Hi again @jgbreezer 

> "manufacturerdata": "0000aabbccddeeff",

With this sample you have given for your GTH+ I assume that this is the only advertising data being broadcast by the watch, or do you ever receive any other data?

Also with the manufacturerdata and you understandably masking out the MAC address, the 0000 would then have been the company ID for the Gateway` "mfr": "Ericsson AB"` recognition.

If this manufacturerdata is all that is being received by the watch I don't really see any point in creating a full decoder for it, as there are no useful encoded properties in the broadcast data, apart from its MAC address, which is already known for this device from being scanned.

So my assumption here is that you would want to use the GTH+ as an auto-discovered device tracker in Home Assistant. Is that the case?

For that you would only want a tracker decoder, best added to the other tracker decoders in

https://github.com/theengs/decoder/blob/development/src/devices/tracker_json.h

> Believe continuous/active scanning may be best (how to tell?

Continuous scanning is only really required for devices like contact or motion sensors, where the broadcast data could happen at any time and only for a very brief moment - this is not the case for the watch.

Active scanning is however an issue to consider here,e specially as you have the broadcast name in the model condition. The vast majority of devices only broadcast their name when being actively scanned. As passive scanning is preferable though, as it does not affect the device's battery at all, it might be better to change the model condition to only look at the data length of the manufacturerdata, looking that it starts with the Ericsson AB company ID and that from index 4 of the manufacturerdata it contains the MAC address.

For a decoder like that, only being used as a tracker decoder the type should also be changed to `16 - TRACK - bluetooth tracker`, and only the `Is NOT Company ID compliant` and `Device compatible with presence tracking` bits should be true.

I hope this clarifies things a bit more.

DigiH:
Hi @jgbreezer 

How are you getting on with the GTH/GTH Pro decoder, or let us know if you have any further queries?

DigiH:
@jgbreezer 

I have added the TicWatch GTH (Pro) to the existing tracker decoders, as form your sample above there doesn't seem to be any additional information, like steps or heart rate, encoded in the broadcast manufacturerdata - only the Ericsson AB company id and the MAC address. Hence also no separate device documentation page and `"type":"TRACK"`, plus the addition of the TicWatch GTH (Pro) to the recognised and discovered Device Trackers in the description text.

Have a look at the PR and let us know if you think there should be any additions or changes.

https://github.com/theengs/decoder/pull/512

PR #512, opened 2024-02-07
Mobvoi TicWatch GTH (Pro) added as tracker

DigiH:
## Checklist:

### DIFF
--- a/docs/devices/devices_template.html
+++ b/docs/devices/devices_template.html
@@ -208,10 +208,8 @@
         </div>
       </header>
     <div class="container">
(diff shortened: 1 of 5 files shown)


jgbreezer:
Thank you for adding it in the trackers @DigiH . I lost motivation a bit and would have come back to it at some point but was working on other things. Learnt enough during it to make it a bit easier to add something else next time though, which there probably will be in due course. Am going to test this branch.

DigiH:
@jgbreezer 

When trying out the branch with Theengs Gateway please make sure that you're also testing it with a current development branch build of Gateway, as there was an issue with discovering device_tracker without any further properties, like the TicWatch now, which was fixed in the last commit in the development branch.

PR #535, opened 2024-04-03
fix: compiler warning: dereferencing type-punned pointer will break strict-aliasing rules

devmirek:
## Description:
During compilation I get gcc warning: decoder.cpp:106:22: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
This fix allows to compile the library with -Wall -Wextra -Werror parameters

## Checklist:


DigiH:
Thanks

PR #598, opened 2025-01-07
Modify Shelly BluH&T to include zero value for button property

chrestme:
## Description:
This modifies the behavior of Shelly BLU H&T sensors so they should always report a value for the button property, even if it isn't reported in the message sent. If the value isn't sent by the sensor, the property is set to zero.

This addresses an issue where downstream applications (Theengs Gateway, BTtoMQTT, Home Assistant) would update the state on the first press, but state wouldn't return to a "non-pressed" state.  

## Checklist:
  - [ ✅ ] The pull request is done against the latest development branch
  - [ ✅ ] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ ✅ ] I accept the [DCO](https://github.com/theengs/decoder/blob/development/docs/participate/development.md#developer-certificate-of-origin).


DigiH:
Thanks

PR #608, opened 2025-02-22
Add support for VCHON VCH6003 Thermo-Hygrometer

johnmave126:
## Description:
Add support for VCHON VCH6003
My units were bought from: https://detail.1688.com/offer/801226352792.html

This is a compact temperature-humidity sensor. From my test, the temperature range is -20-60 C with 0.1C precision, and the humidity range is 20-99% RH with 1% precision.

## Checklist:


DigiH:
Hi @johnmave126 

Thanks a lot for the submission! :)

Please note that all decoder which have "mac@index"/"revmac@index" in the model condition should also have an alternative "conditionnomac" model condition, so that the Theengs App open iOS/iPadOS is also able to recognise the device model without a MAC address, as iOS/iPadOS mask the MAC address with random UUIDS.

This should also allow you to remove the name evaluation from the default condition, also removing the need for the extra test case section you had to create below, but to add the tests under the existing **test_mac_mfgdata** section. Sorry for the extra good work on the additional section.

Additionally you should add an additional 01 to the decoder tag, to have it be 
`tag":"0101",`

for an additional `\"cidc\":false` in the resulting JSON, as from out previous conversation the manufacturerdata company ID seems to be only from the Bluetooth chip company, but not the actual device manufacturer.

Have a look at the following decoder as an example
https://github.com/theengs/decoder/blob/development/src/devices/IBT_4XS_json.h

johnmave126:
Aha, that makes sense. Will change it soon.

Do I need to rebase after the change?

DigiH:
> Do I need to rebase after the change?

No need to rebase. You can just add an additional commit to this PR, or squash your commits to one single one and then force push, but it doesn't really matter, as the merging with Squash and Merge anyway.

DigiH:
Thanks! Looking great now.

@1technophile - ready to merge. Thanks

1technophile:
Thanks

PR #609, opened 2025-03-06
Add Holyiot devices

kickouille:
Hello,

I added Holyiot beacons so that the decoder can also get the battery level.
Generic Beacons start with 4c00 in the manufacturer data and two different Holyiot beacons I have start with 4c000215, so I based the device selection condition on this and placed it before generic beacons in devices.h. Unfortunately I cannot guarantee 0215 would not be used by other vendors.
I did find a better way to identify these devices, as following UUID can be anything.

Battery level is extracted from the service data. I based this on this source, for a close model which also have a button on it. Some models have vibration detection: https://github.com/custom-components/ble_monitor/issues/1183#issuecomment-1564026881.

Best regards,

Yanik


DigiH:
Hi @kickouille 

I'm afraid the. model condition you are using so far is catching any generic Apple iBeacon protocol device, with

`"condition":["manufacturerdata", "=", 50, "index", 0, "4c000215"],`

which we are already covering in the [iBeacon decoder](https://github.com/theengs/decoder/blob/development/src/devices/iBeacon_json.h).

Could you give some samples of the raw MQTT messages with manufacturerdata and servicedata, so we can see if the Holyiot beacons might have a unique iBeacon format UUID, as the [BM2 decoder](https://github.com/theengs/decoder/blob/development/src/devices/BM2_json.h) has for example, so that they might be more uniquely identifiable. Also the servicedata should also be included in the model condition for such a unique identification with then very generic ibecon broadcast.

kickouille:
Hi @DigiH,

You are right, but I thought `4c000215` would be more specific that `4c00`, isn't it? From you experience, what are the chances that other generic iBeacons might also use the second par `0215`?

In any case, these Holyiot beacons are highly configurable, almost anything can be changed, and they can work in Beacon, iBeacon or Eddystone mode:
- MAC address
- UUID
- Minor/Major

By default, UUID will be all zeroed, here is the data I get from the device I have next to me in two different modes:
Beacon:
Manufacturer data: `ffff021500000000000000000000000000000000271b000ec9`
Service data: `4161f71004f685460406040000`

iBeacon:
Manufacturer data: `4c00021500000000000000000000000000000000271b000ec9`
Service data: `4163************0406040000`

This is the model: https://www.aliexpress.com/item/1005003211033416.html?spm=a2g0o.order_detail.order_detail_item.3.764af19cwzaOvA

I'll have to go and search for the other model I have to see if there are similarities we could exploit. They can be both configured the same way, except that the model I'm showing here have vibration detection, while the other hasn't.

Regards,

Yanik

DigiH:
Hi @kickouille 

> From you experience, what are the chances that other generic iBeacons might also use the second par 0215?

Very high ;) as the additional `0215` is part of the iBeacon format, with `02` actually being the iBeacon defining octet and `15` the length of the following bytes, which are also always the same with the iBBeacon format. I might actually submit an extension the iBeacon decoder with this to make it clearer.

https://github.com/theengs/decoder/pull/610

With which project are you using Theengs Decoder? In addition to manufacturerdata and servicedata, does it also send a name with its BLE advertising broadcasts?

The additional servicedata is definitely the p. It could oint to use to differentiate it from other iBeacons in the model condition.

It should also be possible to create decoders for both formats, but maybe best to stick with the iBeacon format for now.

You should also create some test cases for the new decoder, as these are actually causing the failures at the moment with other iBeacon tests now falling into your decoder, and a document .md page with its details.

Depending on which project you use, could you also post some full undecoded MQTT messages with the advertising data included? Obviously you can anonymise the MAC address, but please then also make sure to anonymise it accordingly if it might actually appear in the servbicedata as well.

kickouille:
Hello,

I use your project for proximity detection, the beacon is in the car and garage door opens automatically as soon as I arrive without any other intervention. That works great, even with your generic beacon device, but the only thing I needed in addition is the battery level, which is hex encoded at index 2 of the service data (0x00 to 0x64 for 0% to 100%), so that I know when it's time to change the battery though home assistant.

Acutally, the device as I created it in the pull request works in my use case.

I didn't want to extend your generic iBeacon device as the way battery level is published is probably very specific to Holyiot...

For the other beacon without vibration, this is the JSON content:
```
manufacturerdata: 4c00021500000000000000000000000000000000271b000ac5
name: Holy-IOT
id: 00:11:22:33:44:55
rssi: -87
servicedatauuid: '5242'
servicedata: 41640011223344550406060000
brand: Holyiot
model: iBeacon
model_id: IBEACON
type: BCON
mfid: 4c000215
uuid: '00000000000000000000000000000000'
major: 10011
minor: 10
txpower: -59
```

This is this model: https://www.aliexpress.com/item/1005006378830124.html

And for the beacon with vibration I spoke about in my last post:
```
manufacturerdata: 4c000215cde0598cbc284b89b3786039a1da9efc271b000ec9
name: Holy-IOT2
id: AA:BB:CC:DD:EE:FF
rssi: -67
servicedatauuid: '5242'
servicedata: 4164aabbccddeeff0405040000
brand: Holyiot
model: iBeacon
model_id: IBEACON
type: BCON
mfid: 4c000215
uuid: cde0598cbc284b89b3786039a1da9efc
major: 10011
minor: 14
txpower: -55
batt: 100
```

Yes, they have a name, and it can also be set to anything. Note that for the one with vibration, I did set a random UUID to test.

I anonymised the MAC address as you suggested. MAC is also present in the service data, right after the battery level.
At the end of the manufacturer data, there are these two values on which I have no control: `271b000ac5` and `271b000ec9`.
Could it be something specific to the vendor/chip/model?

To be honest, if you think it's too complicate to add this device to your project, no problem, I can just use my own code, but I wanted to leave a trace, at least, if someone has the same requirement.

On remark/question: in your documentation to [add a decoder](https://decoder.theengs.io/participate/adding-decoders.html):
1. It's not mentioned that the device needs to be added to devices.h
2. It is not explained in which order the matching is done (I presume first match in the definition order of devices.h from my experimentations)
3. I also found references to the devices in decoder.h `enum BLE_ID_NUM`, but my code seems to work without this.

Best regards,

Yanik

DigiH:
> I use your project for proximity detection, the beacon is in the car and garage door opens automatically as soon as I arrive without any other intervention.

Nice use case! 

> To be honest, if you think it's too complicate to add this device to your project, no problem, I can just use my own code, but I wanted to leave a trace, at least, if someone has the same requirement.

No, I don't think it's too complicated at all to add a decoder for the Holyiot beacons :) I just needed to know the full extend of their advertising broadcast details. And it will be best to included a decoder for them which includes as much data as possible.

As I said before, the servicedata also needs to be included in the model condition, to be able to differentiate the Holyiots from other generic iBeacons. So if you change the model condition to

` "condition":["manufacturerdata", "=", 50, "index", 0, "4c000215", "&", "servicedata", "=", 26, "index", 0, "41", "&", "uuid", "index", 0, "5242"],`

This should make the decoder unique for the Holyiots. Adittinally also adding the MAC address property 

```
…
"mac":{
         "decoder":["mac_from_hex_data", "servicedata", 4]
      }
…
```

```
…
"mac":{
         "unit":"string",
         "name":"MAC address"
      }
…
```

Then also change the tag for the Holyiots, as they are not just generic iBeacons any longer. Changing it to

"tag":"1009",

will make it a `type: TRACK - bluetooth tracker` and will also have it auto-discovered as a `Home/Away BLE device tracker` in HA.

You can actually also see this additional data, by turning on 

```
 -padv PUBLISH_ADVDATA, --publish_advdata PUBLISH_ADVDATA
                        Publish advertising and advanced data (1) or not (0)
                        (default: 0)
```

Additionally it might also be interesting to find out which hex data is actually indicating the vibration, and looking at the Aliexpress site there also seems to be temperature, humidity and pressure encoded in the data, possibly the major and minor or the remaining servicedata octets.

DigiH:
> On remark/question: in your documentation to [add a decoder](https://decoder.theengs.io/participate/adding-decoders.html):
> 
> 1. It's not mentioned that the device needs to be added to devices.h

Correct, that should be added.

> 2. It is not explained in which order the matching is done (I presume first match in the definition order of devices.h from my experimentations)

`I presume first match in the definition order of devices.h from my experimentations` - that is the case, so the amended Holyiots decoder needs to be before the generic iBeacon decoder, so as not to fall into that.

> 3. I also found references to the devices in decoder.h `enum BLE_ID_NUM`, but my code seems to work without this.

It also needs to be added there, once a unique model_id for the decoder has also Benn amended, as it currently also still has the generic iBeacons model_id

`"model_id":"IBEACON",`

DigiH:
Hi @kickouille 

Have a look at

https://github.com/theengs/decoder/compare/development...DigiH:decoder:holyiot

for a working decoder for your purpose.

I think that "mfid" and possibly "uuid", "major", "minor" and "txpower" can be removed, unless they contain some decodable data, but I think it might be encrypted - do you know the decryption key?

It should however be fine for your current intended use to also get the battery level for your garage door opener beacon ;)

DigiH:
@kickouille 

For any Holyiot decoder to be fully integrated into Theengs Decoder we should really be able to correctly decoder all their properties. 

With the current state  added hem as BLE device trackers with an added battery level property, which 
will suit your use case just fine.

https://github.com/theengs/decoder/pull/616

kickouille:
Hi @DigiH,

Sorry for staying quite recently.

Thank you for merging the device, I guess the matching conditions are fine, at least in my case.

To answer some of your questions:
- There is no encryption key.
- You are right, some devices from this vendor also provide humidity and temperature, but none of the ones I have do provide this, so I'm unable to say how to extract the data, as it is not documented.
- Yes, one of my devices do provide vibration, I have been able to identify the data, unfortunately this is very erratic (each vibration sets the bit, so it is constantly changing), and it would require some hysterisis control, like the mechanism you use for the tracker, with a timeout. So it will be for another time.
- Device tracking and battery are enough for my usage, even though I would have liked to have "uuid", "major" and "minor".

One thing which could be great, not sure if it requires big changes, is to have the device name in HA corresponding to the name sent by the beacon, instead of the type-PartOfMac. But I guess this is something to be done in the gateway.

Regards,

Yanik

DigiH:
Hi @kickouille 

> One thing which could be great, not sure if it requires big changes, is to have the device name in HA corresponding to the name sent by the beacon, instead of the type-PartOfMac. But I guess this is something to be done in the gateway.

As most BLE devices broadcast static names, we only ever use them for model recognition so far

What is unique to every device, even of the same model, though, is the Bluetooth MAC address/id, from which we also append the last part to the discovery entries. Then any such discovered devices can easily be named in HA.

If, however, you turn on BT: Advanced and Advertising Data, you will see that the name is also being published in the MQTT messages, as visible in your sample messages above. So if you publish a manual discovery message for this name property it will show up alongside the current discovered properties for you.

PR #637, opened 2025-05-17
Fix build system: proper setuptools_scm

1technophile:
## Description:
This is an attempt to fix the wheel building so that Theengs Decoder is available in [piwheels](https://www.piwheels.org/project/theengsdecoder/)
Goal is to support older board like Raspberry Pi zero by providing a ready to install wheel. Building on the zero fail or run for a long time currently.

## Checklist:


DigiH:
Thanks

PR #641, opened 2025-05-23
[CI] Wheel build, install, and import test to PR_build.yml.

1technophile:
## Description:
This commit modifies the `python` job in the `.github/workflows/PR_build.yml` workflow to enhance Python package testing.

The changes include:
- Modifying the build step to create a Python wheel (`bdist_wheel`) instead of a source distribution (`sdist`).
- Adding a step to install the wheel using `pip3`.
- Adding a step to perform an import test (`from TheengsDecoder import decodeBLE`) to verify the package is correctly installed and the core module is accessible.
- Renaming some steps for clarity.

This will help catch packaging and import issues earlier in the PR process.

## Checklist:

### DIFF
--- a/.github/workflows/PR_build.yml
+++ b/.github/workflows/PR_build.yml
@@ -142,16 +142,21 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: '3.11'
-    - name: Install make tools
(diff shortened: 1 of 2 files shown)


DigiH:
Thanks

PR #644, opened 2025-05-26
[SYS] Comment wheel building to fix Pypi publication

1technophile:
## Description:
Keep it only as a test on the PR action.
We may want to add in the future automatic wheel building for multiple architecture but I'm checking which one are relevant.
In the meantime let's unblock the version publication.

## Checklist:


DigiH:
Thanks

PR #655, opened 2025-08-09
Change LYWSD03MMC BTHome from active to just continuous scanning

plambrechtsen:
## Description:

The LYWSD03MMC sensors when running PVVX and BTHome beacons creates two separate frames `LYWSD03MMC_json_PVVX_BTHOME_1` and `LYWSD03MMC_json_PVVX_BTHOME_2` when using OpenMQTTGateway the devices should be configured as continuous scanning and not active scanning as the frames alternate between `BTHOME_1` and `BTHOME_2` sent every 2500ms as per the screenshot below from the PVVX Telink Flasher for Mi Thermostat. What happens is OpenMQTTGateway sometimes sees one frame or the other and creates entitles in Home Assistant as one or the other. When this is swapped over to continuous scanning and Adaptive scan is enabled in OpenMQTTGateway then all the values are correctly created in Home Assistant via OpenMQTTGateway.

image

This only applies to the BTHome beacons as the PVVX and ATC beacons is only a single frame.

## Checklist:

### DIFF
--- a/src/devices/LYWSD03MMC_ENCR_json.h
+++ b/src/devices/LYWSD03MMC_ENCR_json.h
@@ -19,13 +19,13 @@ const char* _LYWSD03MMC_ENCR_json_PVVX = "{\"brand\":\"Xiaomi\",\"model\":\"TH S
    }
 })"""";*/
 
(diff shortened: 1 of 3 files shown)


DigiH:
@plambrechtsen No offence, but continuous scanning for a thermometer is way overkill 😉 
It should only be applied for buttons, motion or contact sensors, where a triggering broadcast could happen at any given moment. Even if you wouldn't mind it personally, I don't think it'd be a good idea too force it upon other adaptive scan users, who might not even use auto-discovery at all.

Now that I'm back home and back in front of a proper large monitor I have created an alternative solution by pulling the two decoders together into one. The existing tests run fine with only the small change of numbering the packets properties to the respective frame.

Can you try out that solution and comment how it is working for you?

https://github.com/theengs/decoder/pull/656

When you proposed this change from active to continuous scanning, did you also set active scanning to continuous, overt does this mean that PVVX-BTHome firmware does send the name with every broadcast so that it can be received with passive scanning?

Could you test the following?

Set `interval` to 20 seconds, keep `scanduration` at 10 seconds, but set `intervalacts` to 5 minutes, and make sure that `pubadvdata` is set to true.

Then monitor the received received messages in MQTT Explorer - are they decoded with the name included every 30 seconds, or only ever 5 minutes?

DigiH:
Closing this, as only continuous scanning without active scanning would not produce any decoded messages.

PR #689, opened 2026-04-26
Rewrite mac@index normalization to avoid UB and a release-build misco…

1technophile:
## Description:
The previous implementation built a std::string from mac_id, then walked it with `for (int x = 0; x < mac_string.length(); x++)` while calling `mac_string.erase(x, 1)` and `mac_string[x] = tolower(mac_string[x])` in the same iteration. Two latent issues:

1. Passing nullptr to std::string's `const char*` constructor (when mac_id is null) is undefined behaviour.

2. Reading `mac_string[x]` immediately after `mac_string.erase(x, 1)` relies on subtle ordering: the erase shifts characters left, so the value at index x changes between the colon check and the tolower read. With aggressive optimization, the compiler is free to cache either the length or the character value across the erase call, which produces a corrupted normalized MAC.

Observed in the wild: TheengsApp built with Android NDK Clang 19 (release pipeline: -O3 + LTO + BOLT) silently returned false from the entire mac@index/revmac@index branch for IBT-2X(S) and MB/SW advertisements, even though the Linux/Python build of the same source decoded them correctly. The pattern above is a plausible trigger.

Rewrite:
- Strip colons + lowercase into a fixed 13-byte stack buffer in a single forward pass over `mac_id`.
- Bail out cleanly if `mac_id` is null or contains fewer than 12 hex chars after stripping (a 12-char comparison cannot match anyway).
- Reuse `cond_len` (12) for the buffer index limit, the reverse length, and the strncmp length, removing the duplicated magic literal.

Verified:
- ctest (BLE + BLE_fail) - 553 + 18 cases, no regressions.
- TheengsApp Android (Qt 6.10.3, NDK 28.2 / Clang 19): IBT-2X(S) and MB/SW vectors that previously decoded as no-match now publish to MQTT correctly.

## Checklist:

### DIFF
--- a/src/decoder.cpp
+++ b/src/decoder.cpp
@@ -274,23 +274,44 @@ bool TheengsDecoder::checkDeviceMatch(const JsonArray& condition,
       } else if (strstr(cond_str, "mac@index") != nullptr) {
         size_t cond_index = condition[++i].as<size_t>();
         size_t cond_len = 12;
-        const char* string_to_compare = nullptr;
-        std::string mac_string = mac_id;
+        char normalized_mac[13];   // 12 hex chars + null terminator
+        char reverse_mac_string[13];
+        const char* string_to_compare = normalized_mac;
 
-        // remove colons and make lower case
-        for (int x = 0; x < mac_string.length(); x++) {
-          if (mac_string[x] == ':') {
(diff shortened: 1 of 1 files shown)


DigiH:
Thanks

PR #690, opened 2026-04-26
iNodeEM: emit batt as int 100 instead of string "100"

1technophile:
## Description:
The `_batt` static-value override was written as `["static_value", "100"]` (quoted), so when the byte-20 condition matches, the property is set to a string `"100"` instead of an int. Other rules' `static_value` numeric literals (e.g. battery percent fallbacks elsewhere) use unquoted numbers, so unwrapping it here aligns with the rest of the decoder and matches what the existing test_ble.cpp `expected` array already declares (`"batt":100`).

Behaviour change is type-only — value is still 100 — but consumers that parse the JSON strictly (e.g. Python json.loads on an MQTT payload) now see an int where the expected output already promised an int. The upstream test passed before only because ArduinoJson 6's JsonVariant `!=` operator silently coerces a numeric string to a number. Strict consumers (Theengs Android app surfacing this via MQTT) saw the string and reported a field-type diff vs the catalog.

## Checklist:


DigiH:
Thanks

PR #698, opened 2026-05-30
feat: build decoder wasm for node and web use

mhaberler:
## Description:

theengs_decoder_wasm.js works with node but fails in a web app like my sensorlogger example

This PR makes it work in both environments.

## Checklist:


mhaberler:
a I see you already addressed this in [add-web-parser-page](https://github.com/theengs/decoder/tree/add-web-parser-page)

PR #708, opened 2026-08-27
add support for victron Orion TR and generic battery monitoring devices

benderl:
## Description:
This PR adds support for the following devices from Victron:
- Orion TR (also called "Orion Smart")
- generic battery monitoring devices, tested with BMV-712 (also called "SmartBMV")

For the SmartBMV a new method `value_from_bit_string` was introduced as the manufacturing data is not byte aligned.

As I am not very familiar with C++, I got lots of help from Copilot. Hope this PR meets your requirements. At least I get meaningful values from my devices.

## Checklist:


DigiH:
Thanks for the contribution @benderl 

I have the following comments
• The Battery Monitor should really be included in VICTRON_SBS_json.h, with the decoder only missing that particular device model version specific identifier "83a3" and the additional properties to the simpler Smart Battery Sense. Could you just amend the already existing decoder and see if it works correctly for you, even without the new `value_from_bit_string` method, as I used masking to decode the non-byte aligned.values.

I suppose we could still introduce this method, at least for this combined decoder, and then change all the relevant Victron decoders to use that instead of the masking at some stage - if you feel like doing that the task ;)

Each device class should only require and have one decoder to keep the size requirement overheads down. With the Battery Monitor and Smart Battery Sense also including the Smart Shunt devices.

• I assume the same issue with possible missing device model version specific identifiers will also apply with the Orion TR, with additional device IDs required to be added to the decoder in the future.

Let me know how you get on with the above.

benderl:
Thanks for your detailed review.

I had the same idea in mind, as Victron uses the same beacon format for each device family. So in theory only one definition per family should be possible.

Lets's see, if this is doable with my limited C++ knowledge and support from Copilot. I think this will take some time.

DigiH:
@benderl 

I took the liberty to amend the Victron SBS decoder to include the Battery Monitor, as well as the SmartShunt, so that all Victron 02 Battery Monitor device types are included in the one single decoder.

~~There are still some condition adjustments necessary for the last few properties~~, as well as including some SmartShunt test case examples – I have some sample messages for a SmartShunt from another user here. I also adjusted the order of the properties to be in line with other Victron devices, as well as adjusting some property names and the property names, which are actually being used to properly define the HA device class sensor types during auto-discovery.

Maybe you could add another test case for your Battery Monitor, and if you are happy with the combined decoder  the still remaining BM decoder can be deleted.

Could you also add separate documentation. pages for the Battery Monitor and Orion TR? Thanks.

Let me know what you think.

benderl:
Hi @DigiH ,
sorry for my late response. The last days were quite busy.
I just tested your changes and still get proper values from my BMV-712.

Regarding the documentation, I integrated the BMV-712 into `VICTRON_SBS.md` and added a new file for the Orion TR.

DigiH:
Thanks

benderl:
I had some more time to validate the official victron docs against the device decoder specifications and found some minor bugs in ORIONTR, SBS and SCC. Should I open separate PRs for each device?

Update: created a PR for each device

